PR #39, opened 2012-08-21
Move user-data and configuration to XDG base directories

Aeyoun:
See commit log for full details.

Complete with updated documentation and even a change log entry.

Should not break anything for any user. The old configuration directory ~/.newsbeuter/ is left in place so any references there will still work.


teto:
Would be great to follow XDG standards. Any update on this ?

gs93:
Move `config` and `urls` to `.config/newsbeuter` and `cache.db`, `history.{search,cmdline}` to `.local/share/newsbeuter`. It's just not the default.

teto:
Could be nice to add a reference to the manpage (I am on 2.8 so maybe it
was added in 2.9)

2015-07-25 17:46 GMT+02:00 Giuliano Schneider <notifications@github.com>:

Minoru:
> Could be nice to add a reference to the manpage (I am on 2.8 so maybe it was added in 2.9)

Good idea! Done. Thanks @gs93 for the comment, too.

@Aeyoun, sorry for taking so long to review this. I just dreaded this PR, it's huge and touches a rather controversial point.

I don't think XDG is a great idea overall, but it certainly has some mindshare, so I'm happy to keep supporting it in Newsbeuter. Not as a default option though, so I'm going to decline this PR.

I tried to cherry-pick your updates to translation file, but it gave me a lot of conflicts caused both by you introducing new strings and Newsbeuter moving on a bit in the last 4 years. I'll be very happy to merge them if you create a separate PR.

Thanks for your work, and sorry it doesn't suit our vision.

PR #159, opened 2015-01-06
Fix broken unit test

ahitrin:
Hello!

I've found that `test.cpp` has become incompilable after the last changes. Please take a look at my attempt to fix this situation.

### DIFF
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -42,7 +42,7 @@ void TestNewsbeuterReload() {
 	lemon.is(feed->items().size(), 8u, "rss.xml contains 8 items");
 
 	rsscache->externalize_rssfeed(feed, false);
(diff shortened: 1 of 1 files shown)


test/test.cpp
```diff
@@ -42,7 +42,7 @@ void TestNewsbeuterReload() {
 	lemon.is(feed->items().size(), 8u, "rss.xml contains 8 items");
 
 	rsscache->externalize_rssfeed(feed, false);
-	rsscache->internalize_rssfeed(feed, NULL);
+	feed = rsscache->internalize_rssfeed("http://testbed.newsbeuter.org/unit-test/rss.xml", NULL);
```
ahitrin: I'm not sure should we reuse `feed` variable here, or it's better to create new variable, like `feed2` below. I've left `feed` variable only because it reduces total size of change.

PR #163, opened 2015-02-18
Fix feed item numbers

jspricke:
Introduced in a3106ff.


akrennmair:
Good catch, thank you!

PR #199, opened 2015-08-23
config: search for ncurses6

mathstuf:


akrennmair:
Since I just merged a very similar pull request that fixes the same issue differently, I'll just close this. In any case, thanks for contributing!

mathstuf:
I don't know if that one would work on Fedora. I'll try it out.

Issue #198, opened 2015-08-16
newsbeuter fails to build with ncurses 6


nilsholland:
newsbeuter (both 2.9, as well as the latest version from git) fails to build / configure on systems with ncurses-6.0 with the following error message:

*Checking for package ncursesw5 using ncursesw5-config... not found*

*You need package ncursesw in order to compile this program.*
*Please make sure it is installed.*

The cause can be found in newsbeuter-2.9/config.sh:
```
if [ `uname -s` = "Darwin" ]; then
        check_custom "ncurses5.4" "ncurses5.4-config" || fail "ncurses5.4"
elif [ `uname -s` != "OpenBSD" ]; then
        check_custom "ncursesw5" "ncursesw5-config" ||  fail "ncursesw"
fi
```
The following patch checks for ncurses 6 and, if not found, falls back to ncurses 5 and only when that version is not found either reports an error condition. Using this patch, newsbeuter seems to build and run just fine on systems using ncurses 6.
```
--- config.sh_orig	2015-08-15 17:03:25.811438366 +0200
+++ config.sh	2015-08-15 17:14:45.107069702 +0200
@@ -110,7 +110,7 @@ check_pkg "stfl" || fail "stfl"
 if [ `uname -s` = "Darwin" ]; then
 	check_custom "ncurses5.4" "ncurses5.4-config" || fail "ncurses5.4"
 elif [ `uname -s` != "OpenBSD" ]; then
-	check_custom "ncursesw5" "ncursesw5-config" ||  fail "ncursesw"
+	check_custom "ncursesw6" "ncursesw6-config" || check_custom "ncursesw5" "ncursesw5-config" || fail "ncursesw"
 fi
 check_ssl_implementation
 all_aboard_the_fail_boat
```

vapier:
it should be using pkg-config instead to locate ncurses settings and not the config scripts at all

Earnestly:
It seems to be looking specifically for version 5 though:
```
Checking for package ncursesw5 using ncursesw5-config... not found
```

But yes, pkg-config would be nicer:
```
teapot earnest %i master ~ pkg-config --libs --cflags ncursesw
-D_GNU_SOURCE -lncursesw
```

akrennmair:
I believe this is fixed now.

Earnestly:
Yup, it appears to build successfully here.  Thank you o/

wimstefan:
Same here ... builds & runs B-)

PR #201, opened 2015-08-27
Use pkg-config to search for ncursesw.

ccxcz:
Newsbeuter doesn't recognise new ncurses version when it's hardcoded to specific config program (issue #198). This checks using pkg-config first.


mathstuf:
Related to #199.

PR #202, opened 2015-09-08
config.sh: check for ncursesw6

Barthalion:


Barthalion:
Actually https://github.com/akrennmair/newsbeuter/pull/201 looks way better than this, so closing.

PR #226, opened 2015-12-19
Add item author to ttrss_api

cvarrick:
Set item.author when using ttrss backenf


Minoru:
I don't have a TT-RSS instance to test it on at the moment, so I'm going to presume you've tested that. (TT-RSS docs are just awesome. Why would anyone want to know what API returns, right?)

Thank you for the pull request!

cvarrick:
Tested with tt-rss version 1.15.3.0c6f7b3, api level 11.

